PR #15, opened 2016-12-01
show pdf preview in the files sidebar

icewind1991:
- [x] requires https://github.com/nextcloud/server/pull/2435

Before:

After:

cc @jancborchardt @MorrisJobke

### DIFF
--- a/js/previewplugin.js
+++ b/js/previewplugin.js
@@ -146,3 +146,38 @@ $(document).ready(function(){
 		}
 	}
 });
+
+(function () {
+
+	var SidebarPreview = function () {	};
+
+	SidebarPreview.prototype = {
+		attach: function (manager) {
+			manager.addPreviewHandler('application/pdf', this.handlePreview.bind(this));
+		},
+
+		handlePreview: function (model, $thumbnailDiv, $thumbnailContainer, fallback) {
+			var previewWidth = Math.floor($thumbnailContainer.parent().width() + 50);  // 50px for negative margins
+			var previewHeight = Math.floor(previewWidth / (16 / 9));
+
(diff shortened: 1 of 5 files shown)


js/previewplugin.js
```diff
@@ -146,3 +146,43 @@ $(document).ready(function(){
 		}
 	}
 });
+
+(function () {
+
+	var SidebarPreview = function () {
+		this.preview = new OCA.Files_Markdown.Preview();
```
nickvergessen: Inter app dependency or mistake?

icewind1991: I would never just copy-past a block of code and forget to properly clean things up

icewind1991:
@nickvergessen fixed

jancborchardt:
Similar here:
```
TypeError: this.preview is undefined in previewplugin.js:153:3
```

Btw when this works, we have another widely used file type ready for a possible grid view, right?

LukasReschke:
> Btw when this works, we have another widely used file type ready for a possible grid view, right?

No. I doubt that this will scale at all. We need a server-side preview generation for that.

MorrisJobke:
Ummm what? We simply load the PDF in the side bar? :)


This will not work for a grid view in my opinion. At least once you have something like 10 PDFs in a folder it will get tricky.

icewind1991:
Fixed the error

I agree that this should be used for grid

MorrisJobke:
> I agree that this should be used for grid

I guess there is a `not` missing ;)

MorrisJobke:
Tested and works 👍 but we should not use this in a grid view.

icewind1991:
> I guess there is a not missing ;)

Correct

jancborchardt:
Okidoke, thanks for clarification :)

Issue #22, opened 2017-02-21
Escape button could quit PDF viewer


jancborchardt:
Similar to how ESC works for quitting the image viewer, it could also work to close the PDF viewer.

andreasjacobsen93:
Did a PR: https://github.com/nextcloud/files_pdfviewer/pull/37

PR #37, opened 2017-06-28
Escape button quits PDF viewer

andreasjacobsen93:
As proposed here https://github.com/nextcloud/files_pdfviewer/issues/22, this PR makes the viewer close on ESC.

### DIFF
--- a/js/previewplugin.js
+++ b/js/previewplugin.js
@@ -44,6 +44,7 @@
 		 */
 		show: function(downloadUrl, isFileList) {
 			var self = this;
+			var shown = true;
 			var $iframe;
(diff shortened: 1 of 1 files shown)


jancborchardt:
Cool @andreasjacobsen93! Please review @nextcloud/javascript :)

js/previewplugin.js
```diff
@@ -73,14 +74,29 @@
...
 						}
 					});
+
+					// Go back on ESC
+					$(document).keyup(function(e) {
+					     if (shown && e.keyCode == 27) // escape key maps to keycode `27`
+					     { 
+					     	shown = false;
+					        if(!$('html').hasClass('ie8')) {
+								history.back();
```
blizzz: Intendation is wrong. But since IE8 isn't supported anymore we could just say self.hide() (also in the block above).

PR #38, opened 2017-07-10
Fix z index for small screen sizes

MorrisJobke:
* open a PDF
* decrease the window width (for smaller than 756 px) the PDF gets overlaid with a white panel
* this fixes it

@karlitschek For this backport - I will also create a master version of this.


karlitschek:
backport makes sense 👍

PR #49, opened 2018-01-16
Fix duplicate iframe id

juliushaertl:
The duplicate id caused issues when closing the pdf preview while the
sidebar was visible. In that case two DOM elements with the same id were
present and the click event handler was registered to the wrong element.

This is a fix for https://github.com/nextcloud/server/issues/4968

Signed-off-by: Julius Härtl <jus@bitgrid.net>


rullzer:
@juliushaertl can you backport to 13 and 12 ?

juliushaertl:
Sure.

stable12: #50 
stable13: #51

PR #106, opened 2018-10-23
adress pdf parameter v2

hirschrobert:
second try
see pr #58 , had trouble with missing sign-off-message

### DIFF
--- a/js/previewplugin.js
+++ b/js/previewplugin.js
@@ -40,14 +40,15 @@
 
 		/**
 		 * @param downloadUrl
+		 * @param param
 		 * @param isFileList
 		 */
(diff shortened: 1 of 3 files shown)


js/previewplugin.js
```diff
@@ -117,7 +118,7 @@
 				actionHandler: function(fileName, context) {
 					var downloadUrl = context.fileList.getDownloadUrl(fileName, context.dir);
 					if (downloadUrl && downloadUrl !== '#') {
-						self.show(downloadUrl, true);
+						self.show(downloadUrl, param, true);
```
ChristophWurst: `param` isn't defined here, this causes a regression

PR #124, opened 2019-03-18
Fix viewing of PDF documents

ChristophWurst:
Fixes a regression of https://github.com/nextcloud/files_pdfviewer/pull/106 that prevents documents from being viewed (clicking a file entry does nothing, error logged to console).

@hirschrobert @skjnldsv please review


rullzer:
And pushed a fix for public previews as well :S

PR #138, opened 2019-06-13
Update pdf.js to v2.0

ChristophWurst:
The goal is to fix rending of Japanese characters in certain documents.

This does not work right now because of two major issues:
- [x] Eval can't be disabled as before, the initialization code fails hard on an CSP violation before we even set the `isEvalSupported` property. A temporary fix to get to the next issue is allowing unsafe eval via the DisplayController. This is not a solution.
- [ ] The worker src is incorrect. No matter if I set `PDFJS.workerSrc` or `PDFJS.GlobalWorkerOptions.workerSrc` it tries to load a file from `/apps/build/...`.

After a few hours of failure I consider myself stuck with this. @danxuliu if you magically have some insights (found some useful comments and commit messages of yours) then please let me know. I'm thankful for any help or suggestions.

---

Sidenote: when I started this migration I figured I could just use the latest version via npm (´pdfjs-dist`) as that is often suggested in the pdf.js repo/wiki. However, it turns out that that is for the library only and not for the viewer …


ChristophWurst:
https://github.com/mozilla/pdf.js/issues/10818 <- CSP issue fixed with the 2.1 beta :raised_hands:

ChristophWurst:
Okay, the 2.x versions do actually not fix the font issues I'm facing. Leaving this PR to the volunteer that wants to complete it :)

PR #176, opened 2020-04-25
Update previewplugin.js

remmedia:
Adds `allow-popups-to-escape-sandbox` to `sandbox` parameter of iframe
Fixes non working links from pdf to other pdf-files an some pages with forms.
Error only occours if externalLinkTarget is set to BLANK


skjnldsv:
Outdated now :)

PR #199, opened 2020-08-17
Fix download prevent, select override and code cleanup

skjnldsv:
- Refactor utils methods for better coding styling
- Fixed hide download compliance
- Fix user-select prevent


rullzer (approved):
:elephant:

PR #201, opened 2020-08-18
Add the framesrc csp by default

rullzer:
Since we want to iframe the pdfviewer we need to properly set it. Else
it might not work on some pages.

Signed-off-by: Roeland Jago Douma <roeland@famdouma.nl>


lib/AppInfo/Application.php
```diff
@@ -27,12 +27,14 @@
 
 namespace OCA\Files_PDFViewer\AppInfo;
 
+use OCA\Files_PDFViewer\Listeners\CSPListener;
```
skjnldsv, line 30: Missing

PR #206, opened 2020-08-19
Fix can download check

skjnldsv:
If there is no element, there is no download to hide

If you are on the files list for example, you cannot download nor print a file because of this. Now you can.


szaimen (approved):
The fix looks good to me 👍

PR #289, opened 2021-01-17
update to pdf.js 2.6.347

salcromic:
update pdf.js: 2.4.456 to 2.6.347

### DIFF
--- a/pdfjs-get.js
+++ b/pdfjs-get.js
@@ -6,7 +6,7 @@ const cliProgress = require('cli-progress')
 
 
 // Fetching pdf.js build release
-const PDFJSversion = '2.4.456'
+const PDFJSversion = '2.6.347'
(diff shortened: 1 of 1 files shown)


skjnldsv:
Hi, It's not as easy, Mozilla changed their release content and our app needs to adapt to it.

skjnldsv:
See https://github.com/nextcloud/files_pdfviewer/pull/225

skjnldsv:
ALso, v2.6.347 is not released as stable yet

salcromic:
Hi, does the update 2.5.207 in Nextcloud support interactive PDFs (edit and save in browser)? This would be very very important to me, I would also like to support this financially.

skjnldsv:
I can't tell. You need to check the changelog of pdf.js https://github.com/mozilla/pdf.js/releases

PR #292, opened 2021-01-22
Only attempt to use a secure view if hide download is actually set

juliushaertl:
Only assume using a secure viewer (Collabora) if hideDownload is actually enabled 

Fixes https://github.com/nextcloud/richdocuments/issues/1320


rullzer (approved):

rullzer:
@juliushaertl backports?

fabianbur:
The current version of PDF viewer is 2.0.1 and it hasn't been updated in a long time. At what time will this fix be included in a version that is updated in the normal way?

Is there a beta version that can be installed until an update is released?

Issue #286, opened 2021-01-15
pdf viewer does not open in a public link share


livier:
### Steps to reproduce
1. Share a pdf file with public link
2. Test the link

### Expected behaviour
Pdf viewer should open

### Actual behaviour
Only see the download page

## Server configuration detail

**Operating system:** Linux 5.4.41-1-pve #1 SMP PVE 5.4.41-1 (Fri, 15 May 2020 15:06:08 +0200) x86_64

**Webserver:** Apache/2.4.25 (Debian) (apache2handler)

**Database:** mysql 10.1.47

**PHP version:** 

7.3.25-1+0~20201130.73+debian9~1.gbp042074
Modules loaded: Core, date, libxml, openssl, pcre, zlib, filter, hash, Reflection, SPL, sodium, session, standard, apache2handler, mysqlnd, PDO, xml, apcu, bcmath, calendar, ctype, curl, dom, mbstring, fileinfo, ftp, gd, gettext, gmp, iconv, imagick, intl, json, exif, mysqli, pdo_mysql, Phar, posix, readline, shmop, SimpleXML, sockets, sysvmsg, sysvsem, sysvshm, tokenizer, wddx, xmlreader, xmlwriter, xsl, zip, Zend OPcache

**Nextcloud version:** 20.0.5 - 20.0.5.2

**Updated from an older Nextcloud/ownCloud or fresh install:** 

**Where did you install Nextcloud from:** unknown

<details><summary>Signing status</summary>

Array
(
)

</details>

<details><summary>List of activated apps</summary>

```
Enabled:
 - accessibility: 1.6.0
 - activity: 2.13.4
 - analytics: 3.2.0
 - apporder: 0.11.0
 - bruteforcesettings: 2.0.1
 - calendar: 2.1.3
 - cloud_federation_api: 1.3.0
 - comments: 1.10.0
 - contacts: 3.4.3
 - contactsinteraction: 1.1.0
 - dashboard: 7.0.0
 - dav: 1.16.2
 - deck: 1.2.3
 - duplicatefinder: 0.0.4
 - encryption: 2.8.1
 - federatedfilesharing: 1.10.2
 - federation: 1.10.1
 - files: 1.15.0
 - files_downloadactivity: 1.9.0
 - files_pdfviewer: 2.0.1
 - files_sharing: 1.12.2
 - files_trashbin: 1.10.1
 - files_versions: 1.13.0
 - files_videoplayer: 1.9.0
[...]

Issue #305, opened 2021-02-05
Cannot open PDF inside shared folders


hmt:
**Describe the bug**
A publicly shared PDF **inside a folder** cannot be opened in the viewer. It says `Invalid or corrupted PDF file`

**To Reproduce**
1. Create folder and add PDF file to it
2. Share folder and create public link
3. Open public link
4. Click on PDF file to view

**Expected behavior**
File can be viewed

**Screenshots**
<img width="361" alt="image" src="https://user-images.githubusercontent.com/1246495/107015052-a96d3e00-679c-11eb-9a92-a72756577c36.png">
File cannot be viewed

** Platform **
everywhere

**Browser log**
```
Uncaught (in promise) Error: Invalid or corrupted PDF file.
    _callee7$/</< app.js:817
    promise callback*_callee7$/< app.js:815
    promise callback*_callee7$ app.js:776
    Babel 8
        tryCatch
        invoke
        method
        asyncGeneratorStep
        _next
        _asyncToGenerator
        _asyncToGenerator
        open
    webViewerOpenFileViaURL app.js:1947
    webViewerInitialized app.js:1917
    promise callback*run app.js:478
    webViewerLoad viewer.js:218
    EventListener.handleEvent* viewer.js:228
    Webpack 3
        __webpack_require__
        <anonymous>
        <anonymous>
app.js:817:16

```

**Additional context**
It is fine to share a single PDF. A single PDF can be opened just fine. This only happens for PDF files that are in a shared folder.
This happened immediately after the 20.0.7 update.
First posted here: https://github.com/nextcloud/viewer/issues/764

RealKelsar:
Hi,

same Problem here, but I notice it tries a Auth Call to our SAML Provider, but that one gets suppresed by the browser because of CSP.

Opening the same PDF in OnlyOffice from the context Menu does work finde and I do not see such a call there.

RealKelsar:
Guess it is the same Issue as in #286

hmt:
Yes, you're correct. Didn't see that bug. But for me this only appeared in 20.0.7

RealKelsar:
Same here.

Hanskes:
Same problem here, since update to 20.0.7

szaimen:
cc @skjnldsv

ksteinb:
Same here

Mehvix:
Same here.

dsiminiuk:
+1

nderambure:
[...]

Issue #207, opened 2020-08-07
Invalid PDF structure on shared files


CamZie:
### Steps to reproduce
1. user shares link of a PDF file
2. opens the PDF shared by link
3. gets error "Invalid or corrupted file | Message: Invalid PDF structure"

### Expected behaviour
shows and can download PDF file through the shared link without any problems

### Actual behaviour
receives error when opening and downloading the PDF file:
![image](https://user-images.githubusercontent.com/16030886/89629631-6b02cf00-d89e-11ea-9994-d9f2378d6e92.png)

The owner of the shared link can access the PDFs so it does not seem to be corrupted. Also re-shared the PDF and still same problem occurs.

### Server configuration

**Operating system:** Debian 9.13 (stretch)

**Web server:** nginx 1.10.3

**Database:** MariaDB 10.1.45

**PHP version:** 7.2

**Nextcloud version:** NC 18.0.6

**Updated from an older Nextcloud/ownCloud or fresh install:** updated from an older version

**Where did you install Nextcloud from:** Nextcloud website

**Signing status:**
<details>
<summary>Signing status</summary>

```
No errors have been found.
```
</details>

**List of activated apps:**
<details>
<summary>App list</summary>

```
Enabled:
  - accessibility: 1.4.0
  - activity: 2.11.0
  - admin_audit: 1.8.0
  - apporder: 0.10.0
  - bruteforcesettings: 1.6.0
  - calendar: 2.0.3
  - cloud_federation_api: 1.1.0
  - comments: 1.8.0
  - contacts: 3.3.0
  - dav: 1.14.0
  - drop_account: 0.2.1
  - encryption: 2.6.0
  - external: 3.5.0
  - federatedfilesharing: 1.8.0
  - federation: 1.8.0
  - files: 1.13.1
  - files_pdfviewer: 1.7.0
  - files_rightclick: 0.15.2
  - files_sharing: 1.10.1
  - files_trashbin: 1.8.0
  - files_videoplayer: 1.7.0
  - logreader: 2.3.0
  - lookup_server_connector: 1.6.0
  - notes: 3.5.1
  - notifications: 2.6.0
  - oauth2: 1.6.0
  - password_policy: 1.8.0
  - photos: 1.0.0
  - provisioning_api: 1.8.0
  - quota_warning: 1.7.0
  - recommendations: 0.6.0
  - serverinfo: 1.8.0
  - settings: 1.0.0
  - sharebymail: 1.8.0
[...]

PR #309, opened 2021-02-10
Fix viewing PDF files with public share links

wankdanker:
In nextcloud/files_pdfviewer#286, users have reported that PDF files
are not viewable via public share links. The issue was introduced in
c992b55 when attempting to encode parts of the URL for files that
have special characters.

This patch uses the URL Web API to deal with the parts of the URL
in a more specific way. Also, the path and files searchParams are
set based on `this.filename` and `this.basename` in the same way
it is done in the View module when setting `davPath`.

This has been tested on private and shared public links on files
with and without special characters.

### DIFF
--- a/src/views/PDFView.vue
+++ b/src/views/PDFView.vue
@@ -38,26 +38,18 @@ export default {
 		},
 
 		encodedDavPath() {
-			const hasScheme = this.davPath.indexOf('://') !== -1
+			const hasScheme = ~this.davPath.indexOf('://')
+			const url = new URL(hasScheme ? this.davPath : 'https://.' + this.davPath)
+			const pathname = url.pathname
 
-			const pathSections = this.davPath.split('/')
+			url.pathname = pathname
 
-			// Ignore scheme and domain in the loop (note that the scheme
-			// delimiter, "//", creates an empty section when split by "/").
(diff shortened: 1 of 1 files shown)


violoncelloCH:
I just applied this patch on my production instance (cherry-picked onto stable20) as I've the same issue as in #305 (also related: #286 and #207)
it looks good and works so far...

gsnendzynskigmbh (approved):

filou77:
We also applied this patch (onto v20.0.7).
It does also fix things for us.
Thanks.

Davka:
we have also patched our installation and it works.

juliushaertl:
Unfortnately this seems the bring back the original issue as then the dav path sections would no longer be escaped properly. I'll have a look into that.

sseide:
Have the same problem that i cannot open PDF files while PDF viewer app is enabled. Accessing via WEBDAV works fine.
Therefore encoding is wrong and not fixed by now in current version.

Nextcloud 20.0.10 / PDF viewer 2.0.1 seems to double encode the url, e.g every path component like
`blah blub` gets encoded to `blah%2520blub` instead of `blah%20blub`

created new ticket here #421 

best regards,
S.Seide

PR #319, opened 2021-02-16
Fix opening pdf files in public folder shares

juliushaertl:
Adapted version of https://github.com/nextcloud/files_pdfviewer/pull/309 to make sure that files in public shared folders open again and the original fix of c992b55 is preserved.

We need to handle urls for shares differently since they are not a proper dav path currently

closes #309


skjnldsv (approved):

violoncelloCH (approved):
tested and works on NC20

Vestarija:
how to apply this fix please?

nderambure:
@Vestarija : if you use Nextcloud in production, just wait for the next stable release that may come soon.

bobpaul:
@Vestarija a sort of ham-fisted "fix" is to just delete the `nextcloud/apps/files_pdfviewer` folder from the server. Without that folder PDFs will just download when clicked without trying the viewer. When you install NextCloud 20.0.8 it'll re-install the folder.

skjnldsv:
@bobpaul you can just disable the app

Lozrus:
@bobpaul You can also set-up OnlyOffice or the "EPUB/CBZ/PDF ebook reader" app, as alternative PDF viewers. I've settled on OnlyOffice for my servers (which is not a bad viewer at all) while we wait for the fix to filter down.

Homer-Sim:
Hello I am on 20.0.9 and his issue is still happening. To deactivate PDF Viewer is not working either. Download is starting but pdf can not be viewer due to corruption issue. File is ok when opening it directly from Server.

Any other options to solve it without waiting for an update?

Thanks

premar:
Hello, currently running 20.0.9 with same issue. I disabled the plugin but we need the function to view the pdfs.

premar:
I fixed the problem on our server. It was not related to pdfviewer app but the server itself. We using external storage, and its broken in the new 20.0.9 version and 21. See this Issue https://github.com/nextcloud/server/issues/26457. I used the fix from https://blog.nuvotex.de/nextcloud-smb-stringbuffer/ to resolve the problem.

Homer-Sim:
> 
> 
> I fixed the problem on our server. It was not related to pdfviewer app but the server itself. We using external storage, and its broken in the new 20.0.9 version and 21. See this Issue [nextcloud/server#26457](https://github.com/nextcloud/server/issues/26457). I used the fix from https://blog.nuvotex.de/nextcloud-smb-stringbuffer/ to resolve the problem.

Thats IT!!!!!!!!!! GREAT !!!!

supr4s:
Well, on some servers that have been upgraded, problem still exists despite the application of the patch seen here: https://blog.nuvotex.de/nextcloud-smb-stringbuffer/

For this server, users have their mount volumes via an SMB share.

Upgrade from Nextcloud 19.0.4 to 20.9 => non-functional. Implementation of the patch => problem still present.

Upgrade to 20.9 green 21.0.1 => still the same.

In addition, another problem: when a user downloads from the web interface a .docsx, .xlsx etc and opens it on his computer => corrupted/damaged file, impossible to recover. And for all downloaded files! 

I had to go back via a snapshot of the server to return to version 19.0.4 which had no problem, and which allows users to work correctly.
 
I will try again in a few weeks depending on the progress of the issues

PR #435, opened 2021-06-24
Update stable22 target versions

nickvergessen:
Signed-off-by: Joas Schilling <coding@schilljs.com>


skjnldsv (approved):

.github/workflows/phpunit.yml
```diff
@@ -20,15 +20,16 @@ jobs:
       matrix:
         php-versions: ['7.4']
         databases: ['sqlite']
-  
-    name: php${{ matrix.php-versions }}-${{ matrix.databases }}
+        server-versions: ['stable22']
```
skjnldsv, line 23: My using matric is forces to change the tests checks for every branch :(

nickvergessen: I guess in this app it doesn't matter. But it becomes a problem if someone does a PR to your PR, because then the base_ref does not exist on server.

Since the "change it on every branch" is automated by nickv-bot I don't mind if you don't

skjnldsv: > I guess in this app it doesn't matter. But it becomes a problem if someone does a PR to your PR, because then the base_ref does not exist on server.

Such a specific use case, it should be fine I think :)
In any case, how do you manage all those branchs checks differences in the github branch protection?
One rule per stable?

nickvergessen: Most apps protect `stable[0-9][0-9]`

PR #463, opened 2021-08-13
Fix body footer hiding

skjnldsv:
For some reason on 23 the footer is alongside the `app-content`, but not on 22, 21 nor 20...


PVince81 (approved):
:+1:

PR #472, opened 2021-08-19
Fix download & print view

skjnldsv:
Restoring the old behaviour but fixing the implementation.
We need to also use the same method for the Viewer (folder public shares)
Since both are using the iframe (single and folder link shares), I decided to pass the param in the URL.
This is anyway a workaround as we can't really disable the download since we display the pdf anyway :shrug:


src/workersrc.js
```diff
@@ -28,6 +28,12 @@ import redirectIfNotIframe from './utils/redirectIfNotIframe'
 // Checks if the page is displayed in an iframe. If not redirect to /.
 redirectIfNotIframe()
 
+// Retrieve the canDOwnload from the url, this is
```
artonge, line 31: Sorry:

```suggestion
// Retrieve the canDownload from the url, this is
```

src/public.js
```diff
@@ -42,7 +49,11 @@ window.addEventListener('DOMContentLoaded', function() {
...
 		const downloadUrl = generateUrl('/s/{token}/download', { token: sharingToken })
-		const viewerUrl = generateUrl('/apps/files_pdfviewer/?file={downloadUrl}#page={page}', { downloadUrl, page })
+		const viewerUrl = generateUrl('/apps/files_pdfviewer/?file={downloadUrl}&canDownload={canDownload}#page={page}', {
+			canDownload: canDownload() ? 1 : 0,
+			downloadUrl,
+			page,
+		})
 
 		// Create viewer frame
 		const viewerNode = document.createElement('iframe')
```
artonge, line 59: Looking at the `<iframe>` specs, maybe the `sandbox="allow-downloads"` attribute could be useful. But I am not sure that it restricts every kind of download:

> allow-downloads: Allows for downloads to occur with a gesture from the user.

If it really does restrict every download, it would be a cleaner way to forward `canDownload`.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe

artonge: Arf, Safari does not support it...

artonge (changes requested):

src/public.js
skjnldsv: Well, there are wider things to manage than downloading. Printing for exampel and selecting the text. So i'd rather have one implementation to rule them all :ring:

src/workersrc.js
skjnldsv: Solved :wink:

artonge (approved):

PR #589, opened 2022-04-08
Fix unknown variable after upgrade from v2.13.216

skjnldsv:
Fix https://github.com/nextcloud/server/issues/31795

Ref https://github.com/mozilla/pdf.js/releases/tag/v2.13.216


PVince81 (approved):
:+1:

PVince81:
can you do a new release ? @skjnldsv 

I guess it will be compatible with older stable versions

skjnldsv:
> can you do a new release ? @skjnldsv
> 
> I guess it will be compatible with older stable versions

This is shipped

PR #611, opened 2022-06-23
Ensure the content takes the whole height when we hide the footer

artonge:
The [`#content` height is limited by the `<footer>` height](https://github.com/nextcloud/server/blob/stable24/core/css/public.scss#L40=), but we hide the `<footer>` when the pdf viewer is used, so we need to manually change the `#content` height to take the whole height.

Before | After
-- | --
Screenshot from 2022-06-23 12-40-46 | Screenshot from 2022-06-23 12-41-01


skjnldsv (approved):

skjnldsv:
Needs test fix and node compile :)

PR #619, opened 2022-06-28
Fix pdfJs template

skjnldsv:
Template got updated and I missed that :see_no_evil: 

From
https://github.com/nextcloud/files_pdfviewer/blob/4daeddd5e06e1c2030db3177a6c64c4940e39b52/js/pdfjs/web/viewer.html#L1-L416


artonge (approved):
So we need to copy past it every time it changes ? Can't we include it or something ?

danxuliu (approved):
Tested and works :+1:

skjnldsv:
> So we need to copy past it every time it changes ? Can't we include it or something ?

Yeah, we customised it a bit.
The maintenance need is not that crazy, didn't want to spend that much time on some kind f automation for now :)

PR #620, opened 2022-06-29
Allow slideshow with PDFs

### LINKED ISSUE #444: Possible to utilize slideshow controls from nextcloud/viewer?
### Steps to reproduce
1. Enable Files viewer and Pdf viewer apps
2. Navigate to folder of PDFs
3. Open PDF in viewer lightbox

### Expected behaviour
(More desired than expected behavior) Utilize same left/right controls that files viewer uses for photos/videos.

### Actual behaviour
No controls, can only open a single PDF at a time.

### Server configuration

**Operating system**:
Debian

**Web server:**
Apache2

**Database:**
MySql 10.3.30

**PHP version:**
7.4.19

**Nextcloud version:** (see Nextcloud admin page)
21.0.2

**Where did you install Nextcloud from:**
cloudamo script

(Excluding the rest of the info, as this is not an issue, but a feature request.)

Is this feature request a possibility? I understand from reading the initial issue submitted for this app to use nextcloud/viewer there may be limitations.

skjnldsv:
Fix #444


artonge (approved):

szaimen (changes requested):
blocking. See my comment above

skjnldsv:
> blocking. See my comment above

Hum, no idea how to solve that. Jan approved adding the slideshow

szaimen:
> Hum, no idea how to solve that. Jan approved adding the slideshow

yeah, but he was probably not aware about the conflict with these buttons...

skjnldsv:
@jancborchardt @nimishavijay what do you reckon ? :)

artonge:
Nice catch @szaimen I would put the navigation behind the menu. I just checked and that's what the Gnome image viewer is doing:

Closed | Open
-- | --
Screenshot from 2022-06-29 11-23-33 | Screenshot from 2022-06-29 11-23-20

nimishavijay:
Suggestion by @artonge sounds good to me 🚀

szaimen:
> Nice catch @szaimen, I would put the navigation behind the menu. I just checked and that's what the Gnome image viewer is doing:

the only problem that I see with that is that the pdfviewer is loaded in an iframe and I am not sure if it is possible to inject the overlay buttons into the iframe...

skjnldsv:
> the only problem that I see with that is that the pdfviewer is loaded in an iframe and I am not sure if it is possible to inject the overlay buttons into the iframe...

Nope, indeed. We'd have to think differently :(
Maybe reduce the arrow's height and hide them when the menu is opened :thinking:

gpz1100:
How about just lowering those bottoms, and/or implementing left/right keys on devices with keyboards?  There's already a handler for the escape key, why not monitor for L/R too?

Actually, for esc, one needs to click on the top bar (where the X is ) first before escape works.

gpz1100:
Has some form of this enhancement made it into the 25.xx beta?  I installed 25.0.0 RC1 into a vm but don't see any change in pdf navigation.  Thank you

skjnldsv:
@gpz1100 no, this is not merged, see assigned 26 milestone

jancborchardt:
> hide them when the menu is opened

For reference, this solution sounds good to me as well. If we can get that event from the PDF viewer we can go ahead with that solution. :)

jancborchardt (commented):
> hide them when the menu is opened

For reference, this solution sounds good to me as well. If we can get that event from the PDF viewer we can go ahead with that solution. :)

Io-Qo:
I think this is a very important feature and should be added soon as navigating through files should be easy within Nextcloud. I reckon the navigation should be added to the sidebar rather than through the slideshow feature, as it would circumvent all problems pointed out by @szaimen and would be an universal approach to all files, not just PDFs etc. Navigation buttons (previous/next file) in the sidebar would make more sense to me.

gpz1100:
I'm confused, is this feature currently offered or will be offered or is being discarded entirely?

PR #655, opened 2022-09-07
fix pdfviewer on public pages

szaimen:
Close https://github.com/nextcloud/server/issues/33871

| Before | After |
|---|---|
| image | image |

Signed-off-by: szaimen <szaimen@e.mail.de>


danxuliu (changes requested):
The bottom of the PDF viewer overflows the window, because [the minimum height of the content is set to 100%](https://github.com/nextcloud/files_pdfviewer/blob/e01afb8ea98a1b6a342c93e5c6edc4b8262adcea/src/public.js#L74). However, with the new style it seems that the minimum height must take into account the height of the header, so it should be `calc(100% - 50px)` (or something similar) instead.

Besides that please note that this may need to be adjusted again once https://github.com/nextcloud/server/issues/33803 is fixed.

szaimen:
> The bottom of the PDF viewer overflows the window, because [the minimum height of the content is set to 100%](https://github.com/nextcloud/files_pdfviewer/blob/e01afb8ea98a1b6a342c93e5c6edc4b8262adcea/src/public.js#L74). However, with the new style it seems that the minimum height must take into account the height of the header, so it should be `calc(100% - 50px)` (or something similar) instead.

done

skjnldsv (approved):

PR #683, opened 2022-11-21
[stable23] Audit fix and pin npm/node engines

PVince81:


nickvergessen:
https://github.com/nextcloud/files_pdfviewer/blob/stable23/package.json#L35-L38

is broken, Needs a patch like https://github.com/nextcloud/activity/pull/1005/commits/3962dc14e1c45d39de532fe7c76ed15538a2caf6

PVince81:
note: I can't compile locally, even with node v14:
```
ERROR in ./src/views/PDFView.vue
Module build failed (from ./node_modules/vue-loader/lib/index.js):
TypeError: Cannot read property 'parseComponent' of undefined
    at parse (/srv/www/htdocs/server/apps-extra/files_pdfviewer/node_modules/@vue/component-compiler-utils/dist/parse.js:15:23)
    at Object.module.exports (/srv/www/htdocs/server/apps-extra/files_pdfviewer/node_modules/vue-loader/lib/index.js:67:22)
 @ ./src/main.js 22:0-38 27:15-22
```

nickvergessen:
Did you run npm audit fix with node 14?

PVince81:
I just did right now and the compilation error is still there.

I'll push it anyway.

PVince81:
alright, so it seems with node v14 even on stable23 it doesn't compile :sob: 

let's see if we can update to v16

nickvergessen:
Otherwise try dropping the lock file and install?😬

PVince81:
nope

also the root cause seems to be a "vue" and "vue-template-compiler" version mismatch, which lead to `@nextcloud/webpack-vue-config`.

I'm trying to upgrade this but now it seems the universe is broken:
```
Fetching pdfjs 2.10.377
 ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ 0% | ETA: 0s | 0/100Error: getaddrinfo ENOTFOUND github.com
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (node:dns:71:26) {
  errno: -3008,
  code: 'ENOTFOUND',
  syscall: 'getaddrinfo',
  hostname: 'github.com'
}
```

will try again later

nickvergessen:
> also the root cause seems to be a "vue" and "vue-template-compiler" version mismatch

Just pin that in the dev dependencies. Works fine usually
https://github.com/nextcloud/spreed/blob/master/package.json#L93

And dependabot updates both in one PR then

https://github.com/nextcloud/spreed/pull/8346

PVince81:
annnnd now it compiles with `make dev-setup && make build-js-production` and it just deletes the "js/*" folder

now, I see the docs say to run `npm run build` but that has the same effect

@skjnldsv can you take over ? this is getting too weird.

I think these are the steps that need to be done:
1. Make node v14 sticky in package.json (not sure if stricly required, but why not)
2. Fix compilation by aligning the "vue" and "vue-template-compiler" versions: `npm install @nextcloud/webpack-vue-config@5`
3. `npm audit fix`
4. manual test
5. commit

skjnldsv:
> * Fix compilation by aligning the "vue" and "vue-template-compiler" versions: `npm install @nextcloud/webpack-vue-config@5`

We do not do major upgrades on stable branches.

skjnldsv (approved):

PVince81:
@skjnldsv how did you manage to compile ?

when I run `npm ci && npm run build` I get this:
```
Error: 

Vue packages version mismatch:

- vue@2.7.13 (/srv/www/htdocs/server/node_modules/vue/dist/vue.runtime.common.js)
- vue-template-compiler@2.6.14 (/srv/www/htdocs/server/apps-extra/files_pdfviewer/node_modules/vue-template-compiler/package.json)
```

PVince81:
oh wait, it's only now that I see that it's taking vue from node_modules of the server repo, I didn't realize that I needed the server to be on stable23 as well for compilation

skjnldsv:
> oh wait, it's only now that I see that it's taking vue from node_modules of the server repo, I didn't realize that I needed the server to be on stable23 as well for compilation

You should not nest them, each need their own node modules

PR #707, opened 2023-02-02
fix: hide back the unwanted buttons

skjnldsv:
Also hide the editor buttons as we do not support this feature (yet)


artonge (commented):
No screenshots ?
Change in build files but not in src ?

skjnldsv:
> No screenshots ? Change in build files but not in src ?

It's a regression. It's putting back the hidden attributes on what was supposed to stay hidden.

PR #767, opened 2023-06-25
[master] Fix npm audit

nextcloud-command:
Auto-generated fix of npm audit


szaimen (approved):
LGTM but didnt test

danxuliu (commented):
UI is broken. However, this seems to be unrelated and to have been caused by https://github.com/nextcloud/files_pdfviewer/pull/762, as rebuilding with node 16 fixes the UI. Therefore, I would merge as is and leave the UI fix for a later pull request.

PR #827, opened 2023-10-08
Create en_GB.json

ostasevych:
Added en_GB localisation


danxuliu (commented):
As explained in #826 the localised strings are provided by the [PDF.js project](https://github.com/mozilla/PDF.js). If you would like to improve the localisation please do it directly in the PDF.js project (or maybe in the Mozilla l10n project, as it seems that [the PDF.js translations are imported from the Mozilla translations](https://github.com/mozilla/pdf.js/blob/8a057dcf90f841a5987c6e860912d7f3e8cce13d/l10n/README.md) rather than being managed by the PDF.js project itself). This will benefit anyone using PDF.js, and of course the localised strings will be eventually included in the PDF viewer of Nextcloud too (once the PDF.js version used by the PDF viewer is updated to a release that includes the improved localisation).

Thanks!

PR #941, opened 2024-03-29
Update uk.js

ostasevych:
Updated Ukrainian localisation considering the conventional internationalisation approach in NC.


susnux:
Thank you very much!
But translation is handled on transifex, so this will be immediately overwritten by the integration.
Instead please join the translators on transifex and adjust the content there:
https://app.transifex.com/nextcloud/nextcloud/files_pdfviewer/

You find the Ukrainian team here:
https://app.transifex.com/nextcloud/teams/64236/uk/

More help and communications with other translators:
https://help.nextcloud.com/c/translations/23

PR #995, opened 2024-07-09
Move build to build-js

AndyScherzinger:
shipped apps during packaging purge the build dir used a packaging-time. Hence we need to move the plugins to a differently named folder.


susnux (approved):

webpack.config.js
```diff
@@ -12,6 +14,15 @@ webpackConfig.entry.public = path.resolve(path.join('src', 'public.js'))
 // keep pdfjs vendor in the js folder
 webpackConfig.output.clean = false
 
+webpackConfig.optimization.minimizer = [new TerserPlugin({
+	extractComments: false,
```
ShGKme, line 18: @susnux Do you remember, why `extractComments: false` here?
Should we have it in the shared config?

susnux: We do not want the comments file to be generated - useless distribution files.
Instead nowadays we want to have proper REUSE license files, see the REUSE Webpack plugin in server for example. Likely can be added to the config but at that time it was more or less just legacy code as we migrated to vite in most cases anyway

PR #938, opened 2024-03-29
Update viewer.php

ostasevych:
Fixing not showing localised strings using Chrome by adding internationalisation tags to `templates/viewer.php` according to the conventional approach in NC to localise apps.

See https://github.com/nextcloud/files_pdfviewer/issues/831

PR #1021, opened 2024-08-22
fix: not showing localised strings using Chrome

susnux:
* Resolves https://github.com/nextcloud/files_pdfviewer/issues/831
* Fixed version of https://github.com/nextcloud/files_pdfviewer/pull/938

Fixing not showing localized strings using Chrome by adding internationalization tags to `templates/viewer.php` according to the conventional approach in NC to localize apps.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,6 +61,6 @@ nbproject
 /tests/karma-coverage
 .php_cs.cache
 node_modules
-/vendor
+**/vendor/
 /build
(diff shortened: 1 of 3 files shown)


skjnldsv (approved):

susnux:
Lets go with the easier fix: https://github.com/nextcloud/files_pdfviewer/pull/1022

PR #1022, opened 2024-08-22
fix: Use the user configured UI language for PDFjs instead of browser language

susnux:
* Closes https://github.com/nextcloud/files_pdfviewer/pull/1021
  *  We do not need to introduce duplicated translations, but just fix the configured language
* Resolves https://github.com/nextcloud/files_pdfviewer/issues/834

PDFjs uses the browser language by default, but users should have consistently the Nextcloud defined language. So now we set the PDFjs language to:
1. The Nextcloud configured language if supported
2. The unflavored version if supported (e.g. `de` instead of `de-DE`)
3. Fall back to browser UI language

We need 3 because if we set an unsupported language PDFjs will default to English, which is worse than the browser UI language.

## Screen recording
My Nextcloud language: Spanish
My Browser UI: German

### Before

[Bildschirmaufnahme_20240822_190219.webm](https://github.com/user-attachments/assets/9929d88b-bc72-464e-abb2-6ed4be670fe4)

### After

[Bildschirmaufnahme_20240822_183724.webm](https://github.com/user-attachments/assets/38327e3a-753c-4b1c-8b85-876ba0a1c656)


skjnldsv (approved):

ostasevych:
Hey, so I cannot understand the solution. How to translate the strings, update them? 

Is transifex supported?

I have modified template file where the strings are localised, plus the Ukrainian version. I can help to migrate main languages to the standard template file, if needed.

After each update I manually correct the template file, add the localiser strings. It is a bit annoying thing. So if no automatised solution found I am thinking just to fork the app.

susnux:
@ostasevych We decided to use the PDFjs translations, but we fixed the language detection, see the closed issue.
Also to improve here @danxuliu is about to upgrade to PDFjs v4 which provides much better translation, even for languages not supported by Nextcloud.

susnux:
Ref: https://github.com/nextcloud/files_pdfviewer/pull/1022

ostasevych:
> @ostasevych We decided to use the PDFjs translations, but we fixed the language detection, see the closed issue.
> Also to improve here @danxuliu is about to upgrade to PDFjs v4 which provides much better translation, even for languages not supported by Nextcloud.

Why? Why not to go in line with nc standards as other apps do? 

Which framework to use in order to improve the translations? How to get know if new strings appear, eg transifex informs on that?

susnux:
> Why? Why not to go in line with nc standards as other apps do?

This app basically wraps the PDFjs library, a lot of strings are set internally.
So not everything could be translated as not every string is inside the template but is loaded from the JS.

> Which framework to use in order to improve the translations? How to get know if new strings appear, eg transifex informs on that?

PDFjs is a mozilla library and uses the Mozilla L10n project for translations, I think all Mozilla products are translated at protoon: https://pontoon.mozilla.org/teams/

susnux:
BTW if you are looking for Ukrainian: It is nearly 100% translated
image

ostasevych:
> BTW if you are looking for Ukrainian: It is nearly 100% translated !
I am a volunteer translator, that is why I am so dissatisfied with the way how this app is built in terms of localisation :(
I feel that I have to continue locally my own way of localising the app with the NC traditional way: just replacing the code manually, adding uk.js and uk.json files to the l10n folder from version to version...

ostasevych:
Don't know where to find those translations for PDFjs

image

susnux:
> Don't know where to find those translations for PDFjs

PDFjs is part of Firefox so the translations are also part of the firefox translation project.

>  I feel that I have to continue locally my own way of localising the app with the NC traditional way: just replacing the code manually, adding uk.js and uk.json files to the l10n folder from version to version...

But this does not fully work, as the strings are not all in the template, but also within the Javascript code of the library.

susnux:
BTW do you experience any translation issues? (Since the merged fix, so with current master branch)

ostasevych:
> BTW do you experience any translation issues? (Since the merged fix, so with current master branch)

Nothing changed with 28.0.9. Instead of Ukrainian I see all strings in English. It is not working completely so has to change the template and add my own translations to my cloud.

ostasevych:
So, I manually replaced `js` subfolder with the one, which is in the main branch, checked and it works.
What I would like to ask you to fix is viewer.css file, as it shows big Cyrillic symbols, so I changed the font size (see my  viewer.css attached upper in the thread).

susnux:
> should I manually download and install the app from master branch?

Just until it is also fixed for 28 and 29 :)

susnux:
> Again, no localisation in 28.0.10. The language is still English.

Yes it is only backported to stable30 and stable29. So you would have to update to at least 29.

ostasevych:
> > Again, no localisation in 28.0.10. The language is still English.
> 
> Yes it is only backported to stable30 and stable29. So you would have to update to at least 29.

What is wrong with backporting to 28.x?..

PR #1026, opened 2024-08-23
[stable30] fix: Use the user configured UI language for PDFjs instead of browser language

susnux:
* Backport of https://github.com/nextcloud/files_pdfviewer/pull/1022

PDFjs uses the browser language by default, but users should have consistently the Nextcloud defined language. So now we set the PDFjs language to:
1. The Nextcloud configured language if supported
2. The unflavored version if supported (e.g. `de` instead of `de-DE`)
3. Fall back to browser UI language

We need 3 because if we set an unsupported lanuage PDFjs will default to English, which is worse than the browser UI language.


szaimen (approved):
🐘

AndyScherzinger (approved):

PR #1112, opened 2025-01-15
fix(composer): fix running composer with `--no-dev` option

szaimen:
Otherwise testing with my easy-test container fails as that uses the --no-dev option for composer install.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
 	},
 	"scripts": {
 		"post-install-cmd": [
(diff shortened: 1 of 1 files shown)


szaimen:
CI failure is unrelated and needs separate fixing...

danxuliu (approved):
Tested and works :+1:

szaimen:
Okay, looks like the problem is not existing in stable30 and below. So backports are not needed.

PR #1113, opened 2025-01-15
feat(CI): add update-workflows.yml workflow

szaimen:
Inspired by https://github.com/nextcloud/.github/pull/244


szaimen:
Looks like this is not going to work as it doesnt without a PAT

PR #1210, opened 2025-05-09
[master] Fix npm audit

nextcloud-command:
# Audit report
No fixable problems found (0 unfixable, 10 only fixable manually using --force)


danxuliu (approved):
Tested and works :+1:

PR #1214, opened 2025-05-20
chore(tests): Cleanup bootstrap.php to be forward-compatible

come-nc:
This uses the new tests/autoload.php from nextcloud/server instead of messing directly with OC private autoloader.

See https://github.com/nextcloud/server/pull/52951 and https://github.com/nextcloud/server/pull/52945 for context.

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -1,22 +1,28 @@
 <?php
 
+declare(strict_types=1);
+
 /**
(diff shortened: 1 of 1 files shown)


Antreesy (approved):
Matches with changes on server/spreed. Looping in @danxuliu as maintainer

danxuliu (approved):
Tests work locally and CI is happy, so :+1:

Issue #833, opened 2023-10-11
Pdf controls not working on mobile browsers


Siepen:
### Steps to reproduce
1. Open pdf file in mobile browser (tested with chrome and brave in several versions)

### Expected behaviour
On top should be the information bar with the filename and close button from nextcloud. Below this should the pdf viewer controls (zoom, print, download....)

### Actual behaviour
There're only the pdf viewer controls and as an overlay you see the filename and close button. The viewer controls can't be used, only the close button in the right corner works. 
Chrome, Edge and Firefox on Windows are working as expected. 

### Server configuration

**Operating system**:
Debian 12
**Web server:**
Apache 2.4.57
**Database:**
MariaDB 10.11.4
**PHP version:**
8.2.7
**Nextcloud version:** (see Nextcloud admin page)
27.1.2
**Where did you install Nextcloud from:**
nextcloud.com

**List of activated apps:**
 - activity: 2.19.0
  - admin_audit: 1.17.0
  - announcementcenter: 6.6.2
  - calendar: 4.5.2
  - circles: 27.0.1
  - cloud_federation_api: 1.10.0
  - comments: 1.17.0
  - contacts: 5.4.2
  - contactsinteraction: 1.8.0
  - dashboard: 7.7.0
  - dav: 1.27.0
  - encryption: 2.15.0
  - event_update_notification: 2.2.0
  - external: 5.2.1
  - federatedfilesharing: 1.17.0
  - federation: 1.17.0
  - files: 1.22.0
  - files_accesscontrol: 1.17.1
  - files_fulltextsearch: 27.0.1
  - files_fulltextsearch_tesseract: 27.0.0
  - files_mindmap: 0.0.29
  - files_pdfviewer: 2.8.0
  - files_reminders: 1.0.0
  - files_rightclick: 1.6.0
  - files_sharing: 1.19.0
  - files_trashbin: 1.17.0
  - files_versions: 1.20.0
  - firstrunwizard: 2.16.0
  - forms: 3.3.1
  - fulltextsearch: 27.0.2
  - fulltextsearch_elasticsearch: 27.0.4
  - groupquota: 0.1.11
  - impersonate: 1.14.0
  - logreader: 2.12.0
  - lookup_server_connector: 1.15.0
  - nextcloud_announcements: 1.16.0
  - notes: 4.8.1
  - notifications: 2.15.0
  - oauth2: 1.15.1
  - onlyoffice: 8.2.0
  - password_policy: 1.17.0
[...]

PR #1222, opened 2025-06-11
Fix header and pdf controls overlap (Issue #833)

pierluigizagaria:
Fixed the header and pdf controls overlap as seen in [#833 ](https://github.com/nextcloud/files_pdfviewer/issues/833)

I couldn't reproduce the issue using Chrome's mobile view in DevTools.
To properly test it, I had to connect a physical smartphone via USB in debug mode and use Chrome's remote debugging feature (chrome://inspect).
Only then was I able to observe the issue and debug.

Before

before

After

after


schnillerman:
Brilliant ❤️ grandeeee

pierluigizagaria:
I've updated the compiled files and force-pushed to keep a single commit

PR #1235, opened 2025-07-20
[master] Fix npm audit

nextcloud-command:
# Audit report
No fixable problems found (0 unfixable, 13 only fixable manually using --force)


szaimen (approved):
LGTM but did not test

PR #1414, opened 2026-04-10
fix: :bug: Update pdf save to use correct dav path and credentials

moodyjmz:
The dav path is legacy when saving a pdf. This becomes an issue with password protected public shares.

We need to provide the request token for public

for user shares, we supply the current user id - the auth etc happens server side via session

To test this change

Place a pdf into a folder. Make this folder public, but password protected and editable.

1. Open using the share link
2. Edit the pdf
3. save it. 
4. It should now save.

@icewind1991 was instrumental in pointing me in the right direction:)


src/utils/davUtils.js
```diff
@@ -4,30 +4,19 @@
...
- */
-export function getUserRoot() {
+	const davUrl = generateRemoteUrl('dav')
 	if (isPublic()) {
-		throw new Error('No user logged in')
+		return davUrl.replace('/remote.php', '/public.php') + `/files/${getSharingToken()}`
 	}
-
-	return `/files/${getCurrentUser()?.uid}`
+	return davUrl + `/files/${getCurrentUser()?.uid}`
```
susnux: This should throw if this ever happens. Because otherwise it would silently resolve to `/files/undefined` and hide the error.

```suggestion
	return davUrl + `/files/${getCurrentUser().uid}`
```

moodyjmz:
@susnux Is this ok now?

danxuliu (changes requested):
Thanks for the fix! I still need to do a proper review, but a couple of points :-)

- When addressing a review and fixing an issue in the original commits please use a `fixup` commit instead and squash the commits once the review is approved. In some cases you might need to rebase and amend the original commit if there are conflicts with later commits or you need to adjust the commit description, but at least in this specific pull request a `fixup` commit would be fine. Of course you can always squash the commits even if they were not originally created as fixup commits.
- Please do not merge the latest main branch into the pull request, but rebase the commits on the latest main branch instead.

Independently of that, once the reviews are approved and the commits squashed in the PDF viewer app you will still need to build the JavaScript files. You can do that by writing a comment in GitHub with `/compile /`, which will trigger the compile bot and automatically add the needed files in its own commit.

If you use `/compile amend /` it will amend the last commit instead, but it is better to use `/compile /` and add the changes in its own commit to ease backports (which can be done also with GitHub comments like `/backport to stable32` or, if you want that our future bot overlords see that you were already polite to them back in the day, `/backport to stable32 please`).

moodyjmz:
> Thanks for the fix! I still need to do a proper review, but a couple of points :-)
> 
> * When addressing a review and fixing an issue in the original commits please use a `fixup` commit instead and squash the commits once the review is approved. In some cases you might need to rebase and amend the original commit if there are conflicts with later commits or you need to adjust the commit description, but at least in this specific pull request a `fixup` commit would be fine. Of course you can always squash the commits even if they were not originally created as fixup commits.
> * Please do not merge the latest main branch into the pull request, but rebase the commits on the latest main branch instead.
> 
> Independently of that, once the reviews are approved and the commits squashed in the PDF viewer app you will still need to build the JavaScript files. You can do that by writing a comment in GitHub with `/compile /`, which will trigger the compile bot and automatically add the needed files in its own commit.
> 
> If you use `/compile amend /` it will amend the last commit instead, but it is better to use `/compile /` and add the changes in its own commit to ease backports (which can be done also with GitHub comments like `/backport to stable32` or, if you want that our future bot overlords see that you were already polite to them back in the day, `/backport to stable32 please`).

Ah, cool, thanks for the pointers, I'll be sure to do that

moodyjmz:
@susnux would you be able to review this?

src/services/uploadPdfFile.js
```diff
@@ -32,18 +31,9 @@ export default async function(filename, data) {
 	const requestConfig = {
 		headers: {
 			'Content-Type': 'application/pdf',
-			// Not needed for public pages, although there is no problem if it
-			// is set.
```
danxuliu, line 36: Is it actually needed for public pages now? At least in a quick test it seems to be working fine even if `requesttoken` is not passed 🤔

danxuliu (commented):
As far as I can see the actual fix is using the `dav` endpoint rather than the `webdav` one (which is the legacy endpoint). Do you mind if I split your commit in a fix replacing the endpoint and a refactor unifying `getRootPath` for future reference? :-)

moodyjmz:
> As far as I can see the actual fix is using the `dav` endpoint rather than the `webdav` one (which is the legacy endpoint). Do you mind if I split your commit in a fix replacing the endpoint and a refactor unifying `getRootPath` for future reference? :-)

sure

src/services/uploadPdfFile.js
moodyjmz: Your domain knowledge is better here than mine, I would have to dive back into the code and refresh myself 😬

danxuliu: Not really... I asked because I do not really know and I did not want to dive into the server code myself :sweat_smile:

moodyjmz: I would need to test this again to see

moodyjmz: @susnux do you want the getRootPath change in this PR but with a separate commit? (I presume for CL)

moodyjmz: @susnux I made that change. I tested:

auth'd user needs the token
public not

previously it was sent for both, I changed it so it is only sent for auth'd users

moodyjmz:
with the recent changes, is that good now?

camilasan:
> with the recent changes, is that good now?

@susnux @danxuliu @szaimen ^^
I need this PR...

src/utils/davUtils.js
danxuliu: Good point, I added an explicit commit for it.

src/services/uploadPdfFile.js
danxuliu: Sorry for the delay. I just restored the comment about `requesttoken` being unneeded and removed it in the last commit which is the one addressing it, and added an explicit commit for what susnux mentioned as it was the case before this pull request.

Other than that, all good, thanks a lot for addressing everything! :-)

danxuliu (approved):
Tested and works 👍

PR #1422, opened 2026-04-17
fix: update CSP

elzody:
The `child-src` directive is deprecated and `ContentSecurityPolicy::addAllowedChildSrcDomain()` [was removed in Nextcloud 34](https://github.com/nextcloud/server/commit/78fd649e475). Instead we can use `addAllowedWorkerSrcDomain()` and `addAllowedFrameDomain()`.

Additionally, the `allowEvalScript` method call was causing errors as well, and it seems like that can just be removed since the app is using the ESM build now.

After these, the app now properly display PDFs for me, whereas there was an internal server error before.

### DIFF
--- a/lib/Controller/DisplayController.php
+++ b/lib/Controller/DisplayController.php
@@ -63,7 +63,7 @@ class DisplayController extends Controller {
 		$response = new TemplateResponse(Application::APP_ID, 'viewer', $params, TemplateResponse::RENDER_AS_BLANK);
 
 		$policy = new ContentSecurityPolicy();
-		$policy->addAllowedChildSrcDomain('\'self\'');
+		$policy->addAllowedWorkerSrcDomain('\'self\'');
 		$policy->addAllowedFontDomain('data:');
 		$policy->addAllowedImageDomain('*');
 		// Needed for the ES5 compatible build of PDF.js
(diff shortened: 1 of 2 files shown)


danxuliu (approved):
Sorry for the delay and thanks for the fix!

There are a couple of details, though :-)
- `frame-src` (`addAllowedFrameDomain`) should not be needed in `DisplayController`; [`frame-src` limits what can be loaded in an iframe](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/frame-src), but the page generated with `DisplayController` is the one loaded in the iframe itself. Therefore `frame-src` should be added on the parent page instead, which [is already done in `CSPListener`](https://github.com/nextcloud/files_pdfviewer/blob/b1808980eb6d9c333a5eb618dd0ee90c545a3dc2/lib/Listeners/CSPListener.php#L26).
- `unsafe-eval` (`allowEvalScript`) allows the use of `eval` in JavaScript, which should be indeed no longer needed, but it also allows compilation and instantiation of WebAssembly (see note in section 5 of https://w3c.github.io/webappsec-csp/#directive-script-src). WebAssembly is needed [when scripting is enabled in the PDF viewer](https://github.com/nextcloud/files_pdfviewer/blob/447e9307cc26a59783be68792bdb61060ff1be4b/README.md#enable-javascript-execution-in-pdf-files) to run the JavaScript sandbox, but it seems it might be needed even without that feature ([specially in PDF.js versions newer than the one that we are currently using](https://redirect.github.com/mozilla/pdf.js/issues/18457#issuecomment-2238653908), which should be updated soon anyway). Therefore, rather than just removing `unsafe-eval` it needs to be replaced by `wasm-unsafe-eval` to still allow WebAssembly.

I took the liberty to amend the commits with those adjustments, I hope it is fine :-)

elzody:
@danxuliu Thanks a lot for the info, did not know all that :)

PR #1434, opened 2026-05-06
chore(performance): make the viewer component async

max-nextcloud:
This reduces the size of the files_pdfviewer-main asset from > 700k to < 100k (uncompressed).

Loading that file on a public share took 180ms for me before.


src/main.js
```diff
@@ -2,7 +2,7 @@
  * SPDX-FileCopyrightText: 2020 Nextcloud GmbH and Nextcloud contributors
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
-import PDFView from './views/PDFView.vue'
+const AsyncPDFViewComponent = () => import('./views/PDFView.vue')
```
ShGKme: Viewer expects the `component` to be a Vue component object, or something with the `name`.

When declared as an anonymous function, `AsyncPDFViewComponent` has no name, and `handler.component` name is just `component`.

ref: https://github.com/nextcloud/viewer/blob/755df35b1a5072801055d5dd4613bc0bcefb6600/src/views/Viewer.vue#L933

You need to either declare it as a named function:

```suggestion
function AsyncPDFViewComponent() {
  return import('./views/PDFView.vue')
}
```

Or use `defineAsyncComponent` wrapper:

```suggestion
import { defineAsyncComponent } from 'vue'

const AsyncPDFViewComponent = defineAsyncComponent(() => import('./views/PDFView.vue'))
```

(But I didn't check if it works when Vue is not shared, but Viewer and the View have different vue packages).

max-nextcloud: > (But I didn't check if it works when Vue is not shared, but Viewer and the View have different vue packages).

Looks like it does not. In both cases this.fileList is empty inside the PDFView component.
I'll see if something like https://github.com/nextcloud/text/blob/main/src/views/ViewerView.js would help.

ShGKme: > Looks like it does not. In both cases

That's unexpected. The first option fixed the name. Only the second option relies on Vue and may break due to the vue import..

max-nextcloud: Viewer tries to inject the Mime mixin into the component: https://github.com/nextcloud/viewer/blob/master/src/views/Viewer.vue#L917
I think that's not working anymore with the async component. Richdocuments works around this by specifying the props on its own and I am also looking in that direction. I don't like mixins and I particularly do not like cross app mixin dependencies.

src/views/ViewerView.js
```diff
@@ -0,0 +1,56 @@
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
```
ShGKme: 🎄

```suggestion
 * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
```

src/views/ViewerView.js
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+import Vue, { defineComponent } from 'vue'
+
+/**
+ *
+ */
+function AsyncPDFViewComponent() {
+	return import('./PDFView.vue')
+}
```
ShGKme: Just to simplify migration to Vue 3

```suggestion
const AsyncPdfView = defineAsyncComponent(() => import('./PDFView.vue'))
```

src/views/ViewerView.js
```diff
@@ -0,0 +1,56 @@
...
+ * based on the options provided here for the options api.
+ *
+ * When mounted this component constructs the vue instance
+ * used inside pdfviewer based on pdvviewers vue import.
+ *
+ * Viewer adds a mixin to the component that adds a bunch of props.
+ * These are all handed down to the PDFView component.
+ */
+export default defineComponent({
+	name: 'ViewerView',
```
ShGKme: Viewer registers components by name. To make sure there is no collision, I'd name more specifically:

```suggestion
	name: 'FilesPdfViewerLoader',
```

src/views/ViewerView.js
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ *
+ */
+function AsyncPDFViewComponent() {
+	return import('./PDFView.vue')
+}
+
+// The vue instance used inside text constructed with the import above.
+let innerVue
```
ShGKme: This variable is in a global scope, but the instance is created in the vue component instance scope. In theory, there could be 2 instances, and creating the second will override the first instance in this variable. (I can imagine something like slider in the viewer, where the next/prev is already loaded).

Instead, it can be stored in the component property ...

src/views/ViewerView.js
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * Viewer adds a mixin to the component that adds a bunch of props.
+ * These are all handed down to the PDFView component.
+ */
+export default defineComponent({
+	name: 'ViewerView',
+	render: (h) => h('div'),
+	inheritAttrs: false,
+	mounted() {
+		innerVue = new Vue({
```
ShGKme: ...

```suggestion
		this.innerVue = new Vue({
```

src/views/ViewerView.js
```diff
@@ -0,0 +1,56 @@
...
+						doneLoading: () => this.doneLoading(),
+					},
+				})
+			},
+		})
+		innerVue.$mount(this.$el)
+	},
+	beforeDestroy() {
+		innerVue.$destroy()
+	},
```
ShGKme:
```suggestion
		this.innerVue.$mount(this.$el)
	},
	beforeDestroy() {
		this.innerVue.$destroy()
	},
```

src/views/ViewerView.js
ShGKme: I'd add a name, may be useful for debugging.


```suggestion
		innerVue = new Vue({
			name: 'FilesPdfViewerProxy',
```

ShGKme (approved):

PR #1438, opened 2026-05-11
chore: update pdfjs-dist to latest v4

szaimen:


danxuliu (changes requested):
Unfortunately it is not as easy as bumping the version ;-) If it was I would have done it long ago :-P

I will (finally) take care of it once I am back, please wait until then :-)

PR #1488, opened 2026-08-05
chore: Bump PHP version to 8.3

danxuliu:
[Nextcloud 35 and later only supports PHP >= 8.3](https://github.com/nextcloud/server/commit/a7da9cd5969c414c66074a3dc4b938e3d6f193ec). Version 8.3.16 had to be explicitly set for psalm, as it is the minimum compatible version with psalm 6.16.1.

This should also fix the [psalm errors in CI](https://github.com/nextcloud/files_pdfviewer/actions/runs/30153587747/job/89667932336) (`Check enforcement of minimum PHP version 8.3 in psalm.xml`).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -4,7 +4,7 @@
 		"optimize-autoloader": true,
 		"classmap-authoritative": true,
 		"platform": {
-			"php": "8.1"
+			"php": "8.3"
 		},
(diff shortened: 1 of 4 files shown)


szaimen (approved):
LGTM but didnt test

PR #1496, opened 2026-08-06
fix: Add explicit dependency for vue-loader@legacy

danxuliu:
Current `vue-loader@legacy` installed as transitive has a vulnerable dependency, and there is no fix for the installed major version. When `npm audit fix` is run it tries to update `vue-loader` to another major, because no version is explicitly specified in the project. Fortunately, [the postinstall script of `webpack-vue-config` verifies that the right version of `vue-loader` is installed](https://github.com/nextcloud-libraries/webpack-vue-config/pull/688) and aborts the process (with a nice warning) if it is not.

However, due to all of the above, [when `npm audit fix` is run by the workflow the command fails](https://github.com/nextcloud/files_pdfviewer/actions/runs/31037509926/job/92602773378#step:6:3225) and no fix is applied (although the error is only visible [when debug logs are enabled](https://github.com/nextcloud-libraries/npm-audit-action/blob/cb7272100b142233ffb0e185e03016fdd82880b7/src/main.ts#L79)). To solve that now an explicit dependency is added again for `vue-loader@legacy`.

Note that [the explicit dependency was removed in the past](https://github.com/nextcloud/files_pdfviewer/pull/1312/changes/efeae4299b4e8d619e4c0c3e16c63af67630e463#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L56) due to being [automatically handled](https://github.com/nextcloud/files_pdfviewer/pull/1301#discussion_r2549743782) when the dependencies are installed, but it is necessary when nothing is installed and the dependency has a vulnerability.

Thanks @ShGKme for the explanation!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -30,7 +30,8 @@
         "@nextcloud/webpack-vue-config": "^7.0.2",
         "adm-zip": "^0.6.0",
         "axios": "^1.13.6",
-        "cli-progress": "^3.12.0"
+        "cli-progress": "^3.12.0",
+        "vue-loader": "^15.11.1"
       },
       "engines": {
         "node": "^24.0.0",
@@ -4695,7 +4696,6 @@
       "integrity": "sha512-97sfH2mYNU+2PzGrmK2haqffDpVASuib9/w2/noxiFi31Z54hW+q3izKQXXQZSNhtiUpAI36uSuYepeBe4wpHQ==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "consolidate": "^0.15.1",
         "hash-sum": "^1.0.2",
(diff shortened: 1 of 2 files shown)


ShGKme (approved):

danxuliu:
> The backport to `stable32` failed. Please do this backport manually.
>
> [...]
> 
> Error: Failed to check for changes with origin/stable32: No changes found in backport branch

I did not realize that the backport to stable32 was not needed, as [vue-loader is already an explicit dependency in that branch](https://github.com/nextcloud/files_pdfviewer/blob/7082c30ab8a3e0c687bac9ff190886d54d50f913/package.json#L58) and therefore [`npm audit fix` works fine](https://github.com/nextcloud/files_pdfviewer/actions/runs/31037509926/job/92602773389).

PR #1502, opened 2026-08-07
chore: Add explicit dependency for terser-webpack-plugin

danxuliu:
`terser-webpack-plugin` was already an implicit dependency, but now it is made explicit to ensure that [the build process will not break if the implicit dependency is removed](https://github.com/nextcloud/files_pdfviewer/actions/runs/31148251240/job/92772257087#step:7:58).

This could happen with an update to webpack >= 5.108.0, as [it replaced `terser-webpack-plugin` with `minimizer-webpack-plugin`](https://github.com/webpack/webpack/commit/e91f5a12e2ec46a69aa6dd8aced05d92a61b2732). However, note that even [if `terser-webpack-plugin` was also replaced with `minimizer-webpack-plugin` in _webpack.config.js_](https://github.com/nextcloud/files_pdfviewer/blob/c0f94ced111b78d33dd432ce8c488814c18d7755/webpack.config.js#L8) in the PDF viewer the build could still break, as [the latest `webpack-vue-config` version (7.0.3) also uses both webpack >= 5.108.0 and terser-webpack-plugin without depending on the later](https://github.com/nextcloud-libraries/webpack-vue-config/pull/775#discussion_r3733851206).


package.json
```diff
@@ -56,6 +56,7 @@
     "adm-zip": "^0.6.0",
     "axios": "^1.13.6",
     "cli-progress": "^3.12.0",
+    "terser-webpack-plugin": "^5.6.1",
```
ShGKme, line 59: This dependency isn't used directly by the `files_pdfviewer` itself. It is used by its dependency - the `@nextcloud/webpack-vue-config`.
- If `@nextcloud/webpack-vue-config` updates, for example, to use `minimizer-webpack-plugin` - it will break again.
- If somebody checks the list of dependencies, it can be found that this one is never imported by the app and accidentally removed

The proper fix should be on the library side:
- https://github.com/nextcloud-libraries/webpack-vue-config/pull/800

ShGKme (changes requested):

package.json
danxuliu: > This dependency isn't used directly by the `files_pdfviewer` itself. It is used by its dependency - the `@nextcloud/webpack-vue-config`.

It is... or what am I missing? :thinking: 
https://github.com/nextcloud/files_pdfviewer/blob/c0f94ced111b78d33dd432ce8c488814c18d7755/webpack.config.js#L8

ShGKme (approved):

package.json
ShGKme: Ahh, I see, correct

PR #1517, opened 2026-08-16
[stable32] Fix npm audit

nextcloud-command:
# Audit report
No fixable problems found (0 unfixable, 22 only fixable manually using --force)


danxuliu:
There are no "real" changes.

PR #1560, opened 2026-09-07
Add a vitest harness for the front-end

solracsf:
The app has no JavaScript tests at all, so nothing under `src/` has ever been covered. This adds vitest plus a `node-test` workflow that follows the same `changes` / `summary` shape as the other npm workflows here, so it can be a required check without getting stuck pending on a pull request that touches nothing relevant.

The two specs it comes with hold on current master. One checks that every viewer parameter `PDFView.vue` reads is actually provided by `templates/viewer.php`; the other that the template still has the same element ids as the pdf.js `viewer.html` it is a copy of. The `check-viewer` job can only tell you that `viewer.html` changed, not that the two have drifted apart.

I have a handful of small fixes ready that each bring their own regression spec, so this one wants to go in first.

Ran `npm run test`, `npm run lint` and `npm run stylelint`.

### DIFF
--- a/.github/workflows/node-test.yml
+++ b/.github/workflows/node-test.yml
@@ -1,0 +1,92 @@
+# SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+# SPDX-License-Identifier: MIT
+
+name: Node tests
+
+on: pull_request
+
+permissions:
+  contents: read
+
+concurrency:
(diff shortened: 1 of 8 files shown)


CarlSchwan (approved):

CarlSchwan:
This already managed to find one regression in pdf.js update PR :)

